Issue #18, opened 2014-08-14
How am I supposed to deal with the last exception?


nicolas-grekas:
I have a code like this:

``` php
$promise->then(function() {
    // ...
    throw new FirstException();

})->then(null, function ($error) {
    // ...
    throw new SecondException();

});
```

`SecondException()` is lost, never handled.
But say for example that I have a PHP error handler that triggers exceptions.
Say then that I am in the middle of my work on the error handling closure (the second one above).
While I code it, I make mistakes, like usual.
How am I supposed to get the SecondException to my eyes so that I can have a great debugging experience? Adding a third "robust" `->then(null, 'debugProvedHandler')`?
I'd prefer having an exception thrown from the last error handler be re-thrown, don't you?
Or any other way I'm missing?

PR #19, opened 2014-08-15
Introduce ExtendedPromiseInterface

jsor:
The ExtendedPromiseInterface extends the PromiseInterface with useful shortcut
and utility methods which are not part of the Promises/A specification.

Implemented methods:
## ExtendedPromiseInterface::done()

**Status**: Implemented

``` php
$promise->done(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null);
```

When working with promises, it's a common issue that exceptions (and/or rejections) get _lost_ because there is no way to break out of a promise chain.

With the new `ExtendedPromiseInterface::done()` you can _end_ a promise chain and unhandled rejections/exceptions will be rethrown in an uncatchable way causing a fatal error.

See:
- [done() vs. then()](https://github.com/reactphp/promise/tree/extended-promise#done-vs-then)
- [ExtendedPromiseInterface::done()](https://github.com/reactphp/promise/tree/extended-promise#extendedpromiseinterfacedone)
- Issue #18 
## ExtendedPromiseInterface::otherwise()

**Status**: Implemented

``` php
$promise->otherwise(callable $onRejected);
```

Shorthand for `$promise->then(null, $onRejected)`.

Additionally, you can type hint the `$reason` argument of `$onRejected` to catch only specific errors.

Example:

``` php
$promise
    ->otherwise(function(\RuntimeException $reason) {
        // Only catch \RuntimeException instances
    })
    ->otherwise(function(\Exception $reason) {
        // Only catch \Exception instances
    })
    ->otherwise(function($reason) {
        // Catch other errors
    });
```

The typehint is checked via reflection from `$onRejected`. Performance impact?

See:
- [ExtendedPromiseInterface::otherwise()](https://github.com/reactphp/promise/tree/extended-promise#extendedpromiseinterfaceotherwise)
## ExtendedPromiseInterface::always()

**Status**: Implemented

``` php
$promise->always(callable $onFulfilledOrRejected);
```

The  `$onFulfilledOrRejected` handler gets called, with no arguments, when the promise either fulfills or rejects.

Example

``` php
$promise->always(function () {
    // close files, connections, etc.
});
```

See:
- [ExtendedPromiseInterface::always()](https://github.com/reactphp/promise/tree/extended-promise#extendedpromiseinterfacealways)
## ExtendedPromiseInterface::progress()

**Status**: Implemented

``` php
$promise->progress(callable $onProgress);
```

Shorthand for `$promise->then(null, null, $onProgress)`.

See:
- [ExtendedPromiseInterface::progress()](https://github.com/reactphp/promise/tree/extended-promise#extendedpromiseinterfaceprogress)

---

I would love to hear feeback
- on the addition of ExtendedPromiseInterface at all
- on the implementations of the methods of ExtendedPromiseInterface

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -189,7 +189,7 @@ function _checkTypehint(callable $callback, $object)
     $expectedException = $parameters[0];
 
     if (!$expectedException->getClass()) {
-        return false;
+        return true;
     }
 
     return $expectedException->getClass()->isInstance($object);
--- a/tests/PromiseTest/PromiseRejectedTestTrait.php
+++ b/tests/PromiseTest/PromiseRejectedTestTrait.php
@@ -310,6 +310,26 @@ trait PromiseRejectedTestTrait
     }
 
     /** @test */
+    public function otherwiseShouldInvokeNonTypeHintedRejectionHandlerIfReasonIsAnExceptionForRejectedPromise()
+    {
+        $adapter = $this->getPromiseTestAdapter();
+
+        $exception = new \Exception();
+
+        $mock = $this->createCallableMock();
+        $mock
+            ->expects($this->once())
+            ->method('__invoke')
+            ->with($this->identicalTo($exception));
+
+        $adapter->reject($exception);
+        $adapter->promise()
+            ->otherwise(function ($reason) use ($mock) {
+                $mock($reason);
+            });
+    }
+
+    /** @test */
     public function otherwiseShouldInvokeRejectionHandlerIfReasonMatchesTypehintForRejectedPromise()
     {
         $adapter = $this->getPromiseTestAdapter();


mtdowling:
After looking at the Promises/A+ spec a bit more in-depth, I see that they just say to use untrusted promises that have a then() method. I'm actually very much in favor of that. You could even check if a promise you're given is first Thennable, then check if it has a then method (that is, if foreign promises need to be treated differently for some reason). This would allow react to interop with any other promise libraries (which would be great because I just wrote one, but it has a different design and abstractions, but still implements the same spec).

jsor:
That was what i proposed [here](https://github.com/reactphp/promise/pull/19#issuecomment-58628919) :)
If there is still a need for a Thenable interface, it could be moved to a separate the repository (i wouldn't go so far and propose it as a PSR ;)).

mtdowling:
Totally! I was wrong! You should just check for a then method. I don't think there needs to be a PSR either (I think the A+ spec is enough). There might not even need to be a Thennable interface really.

jsor:
:+1:

PR #34, opened 2015-07-06
Reject some() (and any()) if the input array contains not enough items

jsor:
This fixes a bug where some() never resolves if the input array contains not enough items.

```php
some([1, 2, 3], 4)->then(function() {
    // Never called
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,6 +515,9 @@ will be the resolution value of the triggering item.
 The returned promise will only reject if *all* items in `$promisesOrValues` are
 rejected. The rejection value will be an array of all rejection reasons.
 
+The returned promise will also reject with a `React\Promise\Exception\RangeException`
+if `$promisesOrValues` contains 0 items.
+
 #### some()
 
 ```php
--- a/README.md
+++ b/README.md
@@ -515,7 +515,7 @@ will be the resolution value of the triggering item.
 The returned promise will only reject if *all* items in `$promisesOrValues` are
 rejected. The rejection value will be an array of all rejection reasons.
 
-The returned promise will also reject with a `React\Promise\Exception\RangeException`
+The returned promise will also reject with a `React\Promise\Exception\LengthException`
 if `$promisesOrValues` contains 0 items.
 
 #### some()
@@ -534,7 +534,7 @@ to resolve (that is, when `(count($promisesOrValues) - $howMany) + 1` items
 reject). The rejection value will be an array of
 `(count($promisesOrValues) - $howMany) + 1` rejection reasons.
 
-The returned promise will also reject with a `React\Promise\Exception\RangeException`
+The returned promise will also reject with a `React\Promise\Exception\LengthException`
 if `$promisesOrValues` contains less items than `$howMany`.
 
 #### map()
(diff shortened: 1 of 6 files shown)


jsor:
ping @cboden @clue @WyriHaximus

tests/FunctionAnyTest.php
```diff
@@ -2,18 +2,38 @@
...
     {
         $mock = $this->createCallableMock();
         $mock
             ->expects($this->once())
             ->method('__invoke')
-            ->with($this->identicalTo(null));
+            ->with(
+                $this->callback(function($exception){
+                    return $exception instanceof RangeException &&
+                           'Input array must contain at least 1 item but contains only 0 items.' === $exception->getMessage();
```
clue: The documentation seems to suggest otherwise. IMO it makes sense to reject this with a `RangeException`, but we should probably also add this special case to the README.

clue:
See remark about documentation for `any()`, otherwise LGTM :+1:

tests/FunctionAnyTest.php
jsor: Added a note the the README (e29472f).

clue:
README now LGTM :)

However, after further consideration, I think this should probably throw a subclass of `LogicException` rather than `RuntimeException`. After all this is in fact a logic error which *could* be detected at "compile time" (i.e. it does not really depend on a runtime decision). What's your view on this?

jsor:
Good point. I've used RangeException after discovering this bug while looking through the [bluebird code](https://github.com/petkaantonov/bluebird/blob/master/src/some.js#L102-L105) and extended the SPL RangeException (which in turn extends RuntimeException).

I think it would be strange to use the name RangeException and extend LogicException because this would differ from the SPL.

Maybe we should use another name...any suggestions?

README.md
```diff
@@ -515,6 +515,9 @@ will be the resolution value of the triggering item.
 The returned promise will only reject if *all* items in `$promisesOrValues` are
 rejected. The rejection value will be an array of all rejection reasons.
 
+The returned promise will also reject with a `React\Promise\Exception\RangeException`
```
joshdifabio: Would `UnderflowException` or `LengthException`, both of which extend `LogicException`, make sense here?

README.md
```diff
@@ -515,6 +515,9 @@ will be the resolution value of the triggering item.
...
+if `$promisesOrValues` contains 0 items.
```
joshdifabio: Would `UnderflowException` or `LengthException`, both of which extend `LogicException`, make sense here?

jsor: LengthException might be an option...

jsor: Renamed to LengthException on cd6bb56.

jsor:

jsor:
Any objections against including this in a 2.4.0 release?

cboden:
Technically isn't this an API break?

jsor:
I think (and iirc we agreed on IRC) that this is a bug fix, because

```php
some([1, 2, 3], 4)
    ->then(function() {
        // Never called
    })
    ->otherwise(function() {
        // Never called
    });
```
simply does nothing. It neither calls the fulfillment nor the rejection handler.

cboden:
ah. :+1: for 2.4 then

Issue #38, opened 2015-07-17
PHP Syntax Error in src/functions.php


NicholasRBowers:
Parse error: syntax error, unexpected '[', expecting ')' in CODE on line 68.

jsor:
This is the short array syntax introduced in PHP 5.4 which React/Promise requires as [minimum version](https://github.com/reactphp/promise/blob/master/composer.json#L9). Looks like you're running PHP 5.3?

jsor:
If you're stuck with PHP 5.3, you can use React/Promise ~1.1 which lately got a [compatibility update](https://github.com/reactphp/promise/releases/tag/v1.1.0).

NicholasRBowers:
I literally just noticed that I didn't change the PHP version on my local machine.  Such a false alarm.

jsor:
No problem :smile:

PR #40, opened 2015-07-17
Fix two PHP syntax errors

NicholasRBowers:
There were two syntax errors in `src/functions.php`.  This pull request fixes them both.

Applies to https://github.com/reactphp/promise/issues/38.


staabm:
as stated in the linked issue this is valid php 5.4 syntax and per composer.json of this lib this is the least version which is supported.

NicholasRBowers:
Yep.  Didn't realize the version of PHP that was running on my local was 5.3.  My bad.

PR #47, opened 2016-02-25
Cancellation handler must only be called once, ignore double cancellation

clue:
As a first step, this ticket aims to document the current behavior. Following, we should fix the first test case and probably also reconsider the second one.

Opening this ticket as an RFC and for the reference.

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -91,7 +91,10 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
             return;
         }
 
-        $this->call($this->canceller);
+        $canceller = $this->canceller;
+        $this->canceller = null;
+
+        $this->call($canceller);
     }
 
     private function resolver(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null)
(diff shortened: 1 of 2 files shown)


jsor:
Thanks for spotting. This should be easy to fix.

```diff
diff --git a/src/Promise.php b/src/Promise.php
index 75379ea..7af9943 100644
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -91,7 +91,10 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
             return;
         }
 
-        $this->call($this->canceller);
+        $canceller = $this->canceller;
+        $this->canceller = null;
+
+        $this->call($canceller);
     }
 
     private function resolver(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null)
```

clue:
> Thanks for spotting. This should be easy to fix.

Yeah, I already prepared the same fix for this issue, but figured I'd bring this up here first and confirm if this is actually the desired behavior. Documentation on cancellation behavior is scarce, also with respect to other promise implementors.

FWIW, I've encountered this issue while implementing #48, to which I've just pushed the same update.

This is now ready for review.

PR #48, opened 2016-02-26
Backport promise cancellation support to v1 Promise API

clue:
This is an independent backport of promise cancellation support for the legacy v1 Promise API. It implements the same API as originally implemented in #20, but does not share a single line of code. Original concept courtesy of @jsor.

The main purpose here is backwards compatibility for promise consumers, which can rely on the cancellation API to be available for newer v2 as well as legacy v1 promises.

### DIFF
--- a/src/React/Promise/Deferred.php
+++ b/src/React/Promise/Deferred.php
@@ -128,11 +128,14 @@ class Deferred implements PromiseInterface, ResolverInterface, PromisorInterface
             return;
         }
 
+        $canceller = $this->canceller;
+        $this->canceller = null;
+
         try {
             $that = $this;
 
             call_user_func(
-                $this->canceller,
+                $canceller,
                 function ($value = null) use ($that) {
                     $that->resolve($value);
                 },
--- a/src/React/Promise/Deferred.php
+++ b/src/React/Promise/Deferred.php
@@ -16,6 +16,15 @@ class Deferred implements PromiseInterface, ResolverInterface, PromisorInterface
 
     public function __construct($canceller = null)
     {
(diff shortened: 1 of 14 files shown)


jsor:
:+1:

clue:
This now also contains the fix from #47 so that the cancellation handler will not be called more than once.

Ready for review :shipit:

src/React/Promise/Deferred.php
```diff
@@ -96,6 +122,35 @@ public function resolver()
...
 
+    public function cancel()
+    {
+        if (null === $this->canceller || null !== $this->completed) {
+            return;
+        }
+
+        $canceller = $this->canceller;
+        $this->canceller = null;
+
```
jsor: For consistency, you should check here for `is_callabe($canceller)` and trigger a notice when `false` like done [here](https://github.com/reactphp/promise/blob/1.0/src/React/Promise/FulfilledPromise.php#L19-L23), [here](https://github.com/reactphp/promise/blob/1.0/src/React/Promise/RejectedPromise.php#L17-L20) and [here](https://github.com/reactphp/promise/blob/1.0/src/React/Promise/Deferred.php#L30-L32).

jsor: Alternatively, you could throw in the constructor, similar to [here](https://github.com/reactphp/promise/blob/1.0/src/React/Promise/Deferred.php#L30-L32).

clue:
Thanks for the hint @jsor, makes perfect sense :+1: I've just pushed an update, I've decided to throw an `InvalidArgumentException` to be more in line with how the `Promise` rejects invalid constructors arguments.

PR #54, opened 2016-04-29
Add PHPUnit requirement to composer config

joshdifabio:


jsor:
Whats's the reasoning behind adding it? I personally use global PHPUnit installations and all other React repositories also don't have it as a dependency.

joshdifabio:
> I personally use global PHPUnit installations and all other React repositories also don't have it as a dependency.
I see. I assumed this was just an oversight to be honest.

> Whats's the reasoning behind adding it?
1) Not all PHP test suites have the same dependencies
2) There are multiple major versions of PHPUnit, and you can't have all of these installed globally
3) What happens if a package introduces an additional requirement for running tests, for example behat? Will this be added to composer.json, but PHPUnit not? Right now you have an unwritten rule that the requirement in order to run tests is PHPUnit (what version?). Why not put this requirement where it belongs, in composer.json?
4) Right now, this workflow doesn't work: `git clone ... ; composer update ; php vendor/bin/phpunit`
5) [Composer](https://github.com/composer/composer/blob/1.1.0-RC/composer.json) do it this way

jsor:
> 1) Not all PHP test suites have the same dependencies

The tests for React/Promise don't have any special requirements (afaik).

> 2) There are multiple major versions of PHPUnit, and you can't have all of these installed globally

Yes, but you can have one with which you run the test suite.

> 3) What happens if a package introduces an additional requirement for running tests, for example behat? Will this be added to composer.json, but PHPUnit not? Right now you have an unwritten rule that the requirement in order to run tests is PHPUnit (what version?). Why not put this requirement where it belongs, in composer.json?

If that happens, we can rethink that.

> 4) Right now, this workflow doesn't work: git clone ... ; composer update ; php vendor/bin/phpunit

But `git clone ... ; composer update ; phpunit` does ;)

> 5) Composer do it this way

I'm not against it, but then we should do it more consequently. Composer also uses the local PHPUnit installation on Travis for example.

joshdifabio:
> If that happens, we can rethink that.

This has already happened, though. It affected me when I tried to contribute. How am I supposed to know which version of PHPUnit you guys use?

joshdifabio:
Thanks for coming back to me on this, even if we see things differently. It's really not very important so I'll close this issue.

jsor:
>> If that happens, we can rethink that.

> This has already happened, though. It affected me when I tried to contribute. How am I supposed to know which version of PHPUnit you guys use?

My point was, that it doesn't matter because the test suite runs with all versions.

clue:
This has come up multiple times already and I'm seeing some use in this and TBH I'm undecided on this. How about we file an issue in the root project (https://github.com/reactphp/react) and then decide this for all member projects?

joshdifabio:
> My point was, that it doesn't matter because the test suite runs with all versions.

How am I supposed to know that?

> The tests for React/Promise don't have any special requirements (afaik).

Or that? (Even you don't seem sure.)

How am I even supposed to know that only PHPUnit is required in order to run the tests?

The answer to all three questions should be that I simply review the composer.json file, or rather just run `composer install`. Except in this case I can't.

clue:
@joshdifabio I think you're making some very valid points, care to file an issue in the root project? :+1:

jsor:
> This has come up multiple times already and I'm seeing some use in this and TBH I'm undecided on this. How about we file an issue in the root project (https://github.com/reactphp/react) and then decide this for all member projects?

👍

PR #61, opened 2016-07-19
Revert automatic cancellation of pending collection promises

jsor:
This reverts automatic cancellation of pending collection promises once the output promise resolves. This was introduced in 42d86b7 and was both unintended and backward incompatible.

**Note:** This is for **2.x**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 CHANGELOG for 2.x
 =================
 
-* 2.4.2 (xxxx-xx-xx)
+* 2.5.0 (xxxx-xx-xx)
 
     * Revert automatic cancellation of pending collection promises once the
       output promise resolves. This was introduced in 42d86b7 (PR #36, released
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ CHANGELOG for 2.x
 * 2.4.2 (xxxx-xx-xx)
 
     * Revert automatic cancellation of pending collection promises once the
-      output promise resolves. This was introduced in 42d86b7 and was both
-      unintended and backward incompatible.
+      output promise resolves. This was introduced in 42d86b7 (PR #36, released
+      in [v2.3.0](https://github.com/reactphp/promise/releases/tag/v2.3.0)) and
+      was both unintended and backward incompatible.
 
 * 2.4.1 (2016-05-03)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG for 2.x
 =================
 
+* 2.4.2 (xxxx-xx-xx)
+
+    * Revert automatic cancellation of pending collection promises once the
+      output promise resolves. This was introduced in 42d86b7 and was both
```
clue: 42d86b7 should probably be replaced with a version tag and/or PR as these are more accessible.

clue:
BC breaks is certainly something we want to avoid and it probably makes sense to revert this :+1: 

I'm unsure if the behavior the original change did introduce is not something others could possibly depend on, thus making this a BC break again :)

Does it make sense to preserve both behaviors through a toggle as originally discussed?

CHANGELOG.md
jsor: This has bee 1 commit from the #36 PR, so i decided to use the commit. Will add the PR and version 👍

jsor:
Good point, i'd like to avoid adding some toggle parameters to the function signatures though.

I'd propose adding notes both to the changelog and the readme to explain the behavior change and how to keep the automatic cancellation. What do you think?

clue:
> Good point, i'd like to avoid adding some toggle parameters to the function signatures though.

I understand this, but I'd like to hear some more thoughts on this. Personally, I've used a very similar cancellation behavior in https://github.com/clue/php-block-react#awaitall. While this may be handy at times, this is not something that *must* be implemented in this promise library.

My position would be: I feel that it may make sense to keep both here, do we have any good reasons against this?

The changelog looks good to me (though the release tag may still be subject to discussion) :+1:

jsor:
> My position would be: I feel that it may make sense to keep both here, do we have any good reasons against this?

I'd rather add a `cancelAll` helper function (like in your [php-block-react](https://github.com/clue/php-block-react/blob/master/src/functions.php#L175-L181)) than a parameter. Thoughts?

> The changelog looks good to me (though the release tag may still be subject to discussion)

Yes, thanks for the hint. Fixed the version. 👍

jsor:
I'd like to make progress on this. Any thoughts on this? Ping @cboden @WyriHaximus @clue

PR #68, opened 2016-10-24
Use done() if available when resolving foreign thenables

jsor:
This adds a check to `Promise\resolve` when resolving foreign Thenables if a `done()` method exists and uses this instead of `then()`.

Ref #66

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -15,6 +15,12 @@ function resolve($promiseOrValue = null)
             $canceller = [$promiseOrValue, 'cancel'];
         }
 
+        if (method_exists($promiseOrValue, 'done')) {
+            return new Promise(function ($resolve, $reject, $notify) use ($promiseOrValue) {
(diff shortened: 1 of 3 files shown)


jsor:
Ping @WyriHaximus @clue @cboden

clue:
Not a big fan of this TBH.

Do any of the known "foreign thenables" actually implement the `done()` method?

I wonder if there's no way we can provide our `done()` logic even if the underlying foreign promise only offers a `then()`method?

jsor:
@clue I haven't found a way to do so. The thrown exceptions are swallowed [here](https://github.com/reactphp/promise/blob/master/src/functions.php#L19).

The only way i know to break out here, would be to use an event loop and throw on nextTick.

jsor:
Closing for too much duck-typing :D.
async-interop/promise (#76) might be a better option for interoperability.

Issue #46, opened 2016-01-22
Enforce \Exception instances as rejection values


jsor:
This also removes support for rejecting with a Promise instance. 

Decide whether the rejection value is still an optional argument  and can be `null`. Enforcing an argument is more inline with `throw` (you can't throw `null`) and let's us remove `UnhandledRejectionException`.

daltones:
+1 :+1:

clue:
:+1:

joshdifabio:
Note that `some()` currently [rejects with an array of promises](https://github.com/reactphp/promise/blob/master/src/functions.php#L99). Maybe we need some kind of composite exception class.

jsor:
Good catch, thanks :+1:

joshdifabio:
I think you should allow `\Throwable` in PHP7. 

```php
__construct($e)
{
  if (PHP_MAJOR_VERSION < 7) {
    if (!$e instanceof \Exception) {
      throw \InvalidArgumentException;
    }
  } elseif (!$e instanceof \Throwable) {
    throw \InvalidArgumentException;
  }
}
```

jsor:
Another good catch :thumbsup:

khelle:
Creating exceptions is very heavy for PHP due to filling them with data on creation and not on throwing. Just check what happens when you create 10k of promises, and reject them all. In best sitatuion you get MemoryOverflow error, in worst core dumped. It is reasonable to allow rejection with different values than Throwable for this purpose. I think the good idea would be to allow string or Throwable or some lazy Throwable object defined inside library.

codedokode:
@joshdifabio

> Note that some() currently rejects with an array of promises. Maybe we need some kind of composite exception class.

Maybe we could solve this if some() would reject with the first exception, and not wait for other values? And make a separate function returning a promise that resolves into an array of rejected values.

I think a rejected value should be something `Throwable` and there were no arrays.

codedokode:
@khelle 

I think you might be mistaking about memory consumption, I ran the following code: 

```
php -r '$e = []; for ($i = 0; $i < 10000; $i++) { $e[] = new Exception("Test"); }; var_dump(memory_get_usage()); '
```

And I get 4-5Mb of used memory (that is 500 bytes per Exception object). Maybe you were using some xdebug options that add additional information into an exception? 

I also tried this code: 

```php
use React\Promise\Deferred;

$p = [];
for ($i=0; $i < 10000; $i++) { 
    $p[] = new Deferred();
[...]

PR #73, opened 2016-11-23
Support foreign thenables in reject()

jsor:
`resolve` supports foreign thenables (introduced in #52), so should `reject`.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -25,7 +25,7 @@ function resolve($promiseOrValue = null)
 
 function reject($promiseOrValue = null)
 {
-    if ($promiseOrValue instanceof PromiseInterface) {
+    if (method_exists($promiseOrValue, 'then')) {
         return resolve($promiseOrValue)->then(function ($value) {
             return new RejectedPromise($value);
         });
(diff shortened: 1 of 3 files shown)


jsor:
Request for review @clue @WyriHaximus @cboden

jsor:
> Note sure how useful this actually is, but I don't really mind despite 👍

It is more about being consistent.

> The instanceof check should be significantly faster and I suppose this may be a function that is called quite often. Does it make sense to OR both conditions and rely on short-circuiting here?

Usually, you will do `reject(new \Exception)` or just `reject()`. So, the `method_exists` will be called anyway in most cases.

The optimization would be to turn the logic around

```php
function reject($promiseOrValue = null)
{
    if (
        !$promiseOrValue ||
        $promiseOrValue instanceof \Throwable ||
        $promiseOrValue instanceof \Exception ||
        !is_object($promiseOrValue) ||
        !method_exists($promiseOrValue, 'then')
    ) {
        return new RejectedPromise($promiseOrValue);
    }

    return resolve($promiseOrValue)->then(function ($value) {
        return new RejectedPromise($value);
    });
}
```

jsor:
Closing as i changed my mind. We probably enforce exceptions as rejection reasons in the future (#46). So, no need to support/encourage usage of thenables as rejection reasons.

PR #76, opened 2016-12-22
Implement Interop\Async\Promise

### LINKED ISSUE #64: Interoperability Implementation
Do you plan to implement https://github.com/async-interop/awaitable? Would you mind to provide feedback?

jsor:
Would be great if anyone from the async-interop team (@trowski, @kelunik, @bwoebi) could have a look at the implementation.

Closes #64

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -105,7 +105,7 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface, 
             $onResolved(null, $value);
         }, function ($reason) use ($onResolved) {
             $onResolved(
-                UnhandledRejectionException::nullOrResolve($reason),
+                UnhandledRejectionException::resolve($reason),
                 null
             );
         });
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -2,7 +2,9 @@
 
 namespace React\Promise;
 
-class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
+use Interop\Async\Promise as AsyncInteropPromise;
+
+class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface, AsyncInteropPromise
 {
     private $canceller;
     private $result;
@@ -95,6 +97,18 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
         $this->canceller = null;
 
         $this->call($canceller);
+    }
+
+    public function when(callable $onResolved)
+    {
+        $this->done(function ($value) use ($onResolved) {
+            $onResolved(null, $value);
+        }, function ($reason) use ($onResolved) {
+            $onResolved(
+                UnhandledRejectionException::nullOrResolve($reason),
+                null
+            );
+        });
     }
 
     private function resolver(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null)
--- a/src/RejectedPromise.php
+++ b/src/RejectedPromise.php
@@ -2,7 +2,9 @@
 
 namespace React\Promise;
 
-class RejectedPromise implements ExtendedPromiseInterface, CancellablePromiseInterface
+use Interop\Async\Promise as AsyncInteropPromise;
+
+class RejectedPromise implements ExtendedPromiseInterface, CancellablePromiseInterface, AsyncInteropPromise
 {
     private $reason;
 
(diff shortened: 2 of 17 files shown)


src/Promise.php
```diff
@@ -97,6 +99,18 @@ public function cancel()
...
+    public function when(callable $onResolved)
+    {
+        $this->done(function ($value) use ($onResolved) {
+            $onResolved(null, $value);
+        }, function ($reason) use ($onResolved) {
+            $onResolved(
+                UnhandledRejectionException::nullOrResolve($reason),
+                null
+            );
+        });
```
kelunik: Both `$onResolved` calls should be wrapped with `try` / `catch` blocks and be forwarded to the error handler, currently the one of `Loop`, once the PR is merged, to the promise error handler: https://github.com/async-interop/promise/pull/19

src/RejectedPromise.php
```diff
@@ -73,4 +75,12 @@ public function progress(callable $onProgress)
     public function cancel()
     {
     }
+
+    public function when(callable $onResolved)
+    {
+        $onResolved(
+            UnhandledRejectionException::nullOrResolve($this->reason),
+            null
+        );
```
kelunik: Same here, needs a `try` / `catch` then.

src/functions.php
```diff
@@ -31,6 +46,14 @@ function reject($promiseOrValue = null)
         });
     }
 
+    if ($promiseOrValue instanceof AsyncInteropPromise) {
+        return new Promise(function ($resolve, $reject) use ($promiseOrValue) {
+            $promiseOrValue->when(function ($reason = null, $value = null) use ($resolve, $reject) {
+                $reject($reason ? $reason : $value);
```
kelunik: Are both `null` here fine?

tests/PromiseTest/AsyncInteropResolvedTestTrait.php
```diff
@@ -0,0 +1,72 @@
...
+        $adapter->promise()->when(function ($e, $v) use (&$invoked, $value) {
+            $this->assertSame(null, $e);
+            $this->assertSame($value, $v);
+            $invoked = true;
+        });
+        $this->assertTrue($invoked);
+    }
+
+    /**
+     * Implementations MAY fail upon resolution with an AsyncInteropPromise,
```
kelunik: If they `MAY` fail, that means we can't rely on it?

jsor: This comment is copied over from @bwoebi's https://github.com/async-interop/awaitable-test/blob/master/src/Test.php#L184 :)

src/functions.php
jsor: I havent't found anything in the specification which defines how the callbacks must be invoked and if none, only `$reason` or both arguments are required.

kelunik: The specification states:

> Any implementation MUST at least provide these two parameters.

But what I was rather asking here is whether `$reject` is fine with `null` being passed?

src/Promise.php
jsor: Yes, but since we're requiring `^0.2.0` and do not require `async-interop/event-loop` there is no other way at the moment.

kelunik: I just fixed my patch, so we can merge the PR soon and have a new release hopefully.

tests/PromiseTest/AsyncInteropResolvedTestTrait.php
kelunik: Oh, right. `fail` in the sense of being `rejected`.

src/functions.php
jsor: Yes, we allow rejection without a reason:
https://github.com/reactphp/promise/blob/2.x/src/functions.php#L26
https://github.com/reactphp/promise/blob/2.x/src/RejectedPromise.php#L9

kelunik: I just wasn't sure because of `UnhandledRejectionException::nullOrResolve`. `RejectedPromise` seems to miss a check for `AsyncInteropPromise` in `__construct` btw.

jsor: > I just wasn't sure because of UnhandledRejectionException::nullOrResolve.

`nullOrResolve` was actually wrong. In case of rejection without a reason, it would call the `when()` callbacks with `null` as first argument signalling success instead of failure. This is fixed in 8073b04.

> RejectedPromise seems to miss a check for AsyncInteropPromise in __construct btw.

Good catch, fixed in 8c68ff6

jsor:
Superseded by #78

PR #77, opened 2016-12-22
Preserve the order of the array with all and map functions

### LINKED ISSUE #74: \React\Promise\all not preserving array key order
Hi I ran into an strange issue and am wondering is this intended behavior?

Basically I'm expecting that for the included code the output would be:

```
.Array
(
    [0] => 1
    [1] => 2
    [2] => 3
)
```

But instead I'm getting the array keys unordered:
```
.Array
(
    [0] => 1
    [2] => 3
    [1] => 2
)
```

Which causes problems for example when json_encoding (PHP thinks it should be encoded as an object not an array).

Code to reproduce:

```php
$loop = Factory::create();

$testData = [
    new Promise(
        function ($resolve) use ($loop) {
            $loop->nextTick(
                function () use ($resolve) {
                    $resolve(1);
                }
            );
        }
    ),
    new Promise(
        function ($resolve) use ($loop) {
            $loop->nextTick(
                function () use ($resolve, $loop) {
                    \React\Promise\resolve()->then( function() use ($resolve, $loop) {
                        $loop->nextTick(function () use ($resolve) {
                            $resolve(2);
                        });
                    });
                }
            );
        }
    ),
    new Promise(
        function ($resolve) use ($loop) {
            $loop->nextTick(
                function () use ($resolve) {
                    $resolve(3);
                }
            );
        }
    )
];


\React\Promise\all($testData)->then(function ($result) {
    print_r($result);
});

$loop->run();
``` 

Thanks!

lordthorzonus:
fixes #74 

I added the changelog entry for the 2.5.x heading (not sure if this is correct).

### DIFF
--- a/tests/FunctionAllTest.php
+++ b/tests/FunctionAllTest.php
@@ -94,4 +94,25 @@ class FunctionAllTest extends TestCase
         all(resolve(1))
             ->then($mock);
     }
+
+    /** @test */
+    public function shouldPreserveTheOrderOfArrayWhenResolvingAsyncPromises()
+    {
+        $queue = new \SplQueue();
+        $asyncPromise = new Promise(function ($resolve) use ($queue) {
+            $queue->enqueue(function () use ($resolve, $queue) {
+                $resolve(2);
+            });
+        });
+
+        $result = null;
+
+        all([resolve(1), $asyncPromise, resolve(3)])->then(function ($resolvedValues) use (&$result) {
+            $result = $resolvedValues;
+        });
+
+        call_user_func($queue->dequeue());
+
+        $this->assertSame([1, 2, 3], $result);
+    }
 }
--- a/tests/FunctionAllTest.php
+++ b/tests/FunctionAllTest.php
@@ -98,21 +98,17 @@ class FunctionAllTest extends TestCase
     /** @test */
     public function shouldPreserveTheOrderOfArrayWhenResolvingAsyncPromises()
     {
-        $queue = new \SplQueue();
-        $asyncPromise = new Promise(function ($resolve) use ($queue) {
-            $queue->enqueue(function () use ($resolve, $queue) {
-                $resolve(2);
-            });
-        });
+        $mock = $this->createCallableMock();
+        $mock
+            ->expects($this->once())
+            ->method('__invoke')
+            ->with($this->identicalTo([1, 2, 3]));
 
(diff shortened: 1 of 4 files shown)


WyriHaximus (approved):

tests/FunctionAllTest.php
```diff
@@ -94,4 +94,25 @@ public function shouldResolveToEmptyArrayWhenInputPromiseDoesNotResolveToArray()
...
+        $result = null;
+
+        all([resolve(1), $asyncPromise, resolve(3)])->then(function ($resolvedValues) use (&$result) {
+            $result = $resolvedValues;
+        });
+
+        call_user_func($queue->dequeue());
+
+        $this->assertSame([1, 2, 3], $result);
+    }
```
jsor: I'd like to see the style of the test matching the rest of the tests. In this case this would something like

```php
/** @test */
public function shouldPreserveTheOrderOfArrayWhenResolvingAsyncPromises()
{
    $deferred = new Deferred();

    $mock = $this->createCallableMock();
    $mock
        ->expects($this->once())
        ->method('__invoke')
        ->with($this->identicalTo([1, 2, 3]));

    all([1, $deferred->promise(), 3])
        ->then($mock);

    $deferred->resolve(2);
}
```

tests/FunctionMapTest.php
```diff
@@ -109,6 +109,30 @@ public function shouldResolveToEmptyArrayWhenInputPromiseDoesNotResolveToArray()
...
+        )->then(function ($resolvedValues) use (&$result) {
+            $result = $resolvedValues;
+        });
+
+        call_user_func($queue->dequeue());
+
+        $this->assertSame([2 ,4 ,6], $result);
+    }
+
+    /** @test */
```
jsor: I'd like to see the style of the test matching the rest of the tests (see my comment on the `all` test).

jsor (changes requested):

clue (approved):

lordthorzonus:
@jsor Unified the test style. Sorry about that :).

jsor (approved):

jsor:
Awesome, thanks a lot @lordthorzonus!

jsor:
Released in [v2.5.0](https://github.com/reactphp/promise/releases/tag/v2.5.0).

PR #81, opened 2017-01-08
Remove race()

### LINKED ISSUE #57: Race() accepts an empty array
`race()` currently accepts an empty array, which seems to be a bug.

https://github.com/reactphp/promise/commit/e8c20be73335581a1069cdfb6113a946bb15a250#commitcomment-17349444

jsor:
See #57 for the discussion.

Closes #57


clue:
Let me quote from the linked issue:

> Giving I have no current use for `race()`, I'm undecided.
Does removing it altogether really resolve the issue here? :)
My main concern here would be lack of documentation, but I'm good with either solution :+1: 

I'm currently undecided, but I'm *okay* with either.

jsor:
A little bit more context: Our current implementation is **not** compatible with the ES6 specification because it does not return a forever pending promise if an empty array is given (currently, it returns a resolved promises with no value).

The discussion in #57 has been whether to remove it completely or make it ES6 compliant.

Given that `race()` has not been proven useful and even bluebird (apart from the native Promises, the defacto standard implementation of promises in JS) recommends against using it (http://bluebirdjs.com/docs/api/promise.race.html), this PR proposes to remove it.

clue:
> The discussion in #57 has been whether to remove it completely or make it ES6 compliant.

IMHO consistency with ES6 *makes sense*, but it's not really a necessity because interoperability doesn't really apply across language boundaries anyway. Hence my vote to first *document* the current behavior and how this may differ from other implementations (afaict there's no mention of ES6 promises in the docs in the first place?).

jsor:
@clue Consistency with ES6 in this case does not mean being consistent just for the sake of it but just to behave correctly.

Our [documentation](https://github.com/reactphp/promise#race) states:

> Returns a promise which is resolved in the same way the first settled promise resolves.

So, the current implementation simply does not behave how we document it. It is arguable whether the return value should be a forever pending promise or if we should return an error (like we do with `any()`  which rejects with a `LengthException` when the array is empty). But it should definitely **not** return a resolved promise which resolves with `null`.

jsor:
I've opened an alternative PR which changes the behavior to return a forever pending promise:  #83

jsor:
Closed by #83

PR #82, opened 2017-01-10
Replace the queue() function with a static Queue class

jsor:
We allow swapping out the queue implementation and this must be done in a bootstrap file before executing anything from react/promise.

Consider the case, were a user defines a bootstrap file in the composer.json were the queue implementation is changed.

```json
{
    "autoload": {
        "files": ["src/bootstrap.php"]
    }
}
```

The `bootstrap.php` contains the code for swapping out the queue implementation.

```php
queue(MyQueueImplementation());
```

Depending on the order in `vendor/composer/autoload_files.php`, our `functions.php` might be loaded **after** the `bootstrap.php` thus making the `queue()` function unavailable to the bootstrap file.

This patch replaces the `queue()` function by a static class which is autoloadable.

The only difference is in the API, the behavior is unchanged (note, that this feature is not released yet).

Enqueuing a task:

```php
// Old
queue()->enqueue($task);

// New
Queue::enqueue($task)
```

Swapping out the queue implementation:

```php
// Old
queue(MyQueueImplementation());

// New
Queue::setDriver(MyQueueImplementation());
```

### DIFF
--- a/src/FulfilledPromise.php
+++ b/src/FulfilledPromise.php
@@ -22,7 +22,7 @@ final class FulfilledPromise implements PromiseInterface
         }
 
         return new Promise(function (callable $resolve, callable $reject) use ($onFulfilled) {
-            queue()->enqueue(function () use ($resolve, $reject, $onFulfilled) {
+            Queue::enqueue(function () use ($resolve, $reject, $onFulfilled) {
                 try {
                     $resolve($onFulfilled($this->value));
                 } catch (\Throwable $exception) {
@@ -40,7 +40,7 @@ final class FulfilledPromise implements PromiseInterface
             return;
         }
 
-        queue()->enqueue(function () use ($onFulfilled) {
+        Queue::enqueue(function () use ($onFulfilled) {
             $result = $onFulfilled($this->value);
 
             if ($result instanceof PromiseInterface) {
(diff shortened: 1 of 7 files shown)


jsor:
> This whole feature is currently undocumented, so I'm not sure how either the existing version or the updated version is preferable from a consumer perspective?

I'm not sure what you mean here. The only difference is the API, behavior is unchanged. I've update the PR description, probably its clearer now what changed.

> What do you think? Should we document this in our README or is this merely internal?

Good point, at the moment it is internal. The overall behavior is unchanged to v2. The original motivation for adding a queue has been to reduce recursion and avoid blowing up the call stack (#22).

A side effect is that you can swap out the queue implementation to make react/promise guarantee future-turn resolution (#4), eg. required by the [promises/a+](https://promisesaplus.com/#point-67) spec.

But this is currently undocumented and not tested.

jsor:
Ping @clue

jsor:
@clue Just to be clear, this feature **is tested**.

> A side effect is that you can swap out the queue implementation to make react/promise guarantee future-turn resolution (#4), eg. required by the promises/a+ spec.
>
> But this is currently undocumented and not tested.

This is untested, not documented and currently considered an internal feature.

I'm going to make this clearer by marking everything related `@internal` in a follow-up PR.

PR #84, opened 2017-01-17
phpdoc for some public methods and functions

kyptov:
I added phpdoc to some public methods and functions. It helps IDE to autocomplete methods.

### DIFF
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -14,6 +14,9 @@ final class Deferred implements PromisorInterface
         $this->canceller = $canceller;
     }
 
+    /**
+     * @inheritdoc
+     */
     public function promise()
     {
         if (null === $this->promise) {
--- a/src/functions.php
+++ b/src/functions.php
@@ -2,6 +2,10 @@
 
 namespace React\Promise;
 
+/**
+ * @param mixed|null $promiseOrValue
+ * @return PromiseInterface
+ */
 function resolve($promiseOrValue = null)
 {
     if ($promiseOrValue instanceof PromiseInterface) {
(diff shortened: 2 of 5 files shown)


src/Deferred.php
```diff
@@ -14,6 +14,9 @@ public function __construct(callable $canceller = null)
         $this->canceller = $canceller;
     }
 
+    /**
+     * @inheritdoc
+     */
```
clue: These look pretty much useless and I suppose if your IDE really depends on these, you may want to switch to another one :-)

src/functions.php
```diff
@@ -2,6 +2,10 @@
 
 namespace React\Promise;
 
+/**
+ * @param mixed|null $promiseOrValue
+ * @return PromiseInterface
+ */
```
clue: I think it makes sense to include docblocks on our public API. However, if we do so, we should *carefully* follow what's documented in our README.

src/Deferred.php
jsor: Yes, standalone inheritdoc is pretty useless and as the [FIG proposal states](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#62-using-the-inheritdoc-inline-tag-to-augment-a-description), it should only be used to augment the description of the parent element.

src/functions.php
```diff
@@ -67,6 +87,11 @@ function any(array $promisesOrValues)
         });
 }
 
+/**
+ * @param array $promisesOrValues
+ * @param $howMany
```
jsor: This is the only place in react/promise where a type (`int`) makes sense because it is not `mixed` and not provided in the function/method signature.

src/Deferred.php
kyptov: I use PHPStorm and with inherit doc it works as expected

jsor: Yes, but it also works without it.

kyptov: Oh, you right. I have a specific project where most of functions/methods don`t contain body (for security reason). So phpdoc was the only way to view returned value. Sorry for bothering you, I am going to close this PR.

PR #94, opened 2017-03-17
Fix circular references when resolving with a promise which follows itself

jsor:
**Note:** this patch is for **2.x**

This rounds up #71 preventing circular references not only when promises are resolved directly with itself but also when promises are resolved with promises which follow itself.

Currently, the following ends in an infinite loop:

```php
$d1 = new \React\Promise\Deferred();
$d2 = new \React\Promise\Deferred();

$d1->resolve($d2->promise());
$d2->resolve($d1->promise());

$d2->promise()
    ->otherwise(function(\LogicException $e) {
        assert('Cannot resolve a promise with itself.' === $e->getMessage());
    })
;
```

See also the [Travis build](https://travis-ci.org/reactphp/promise/builds/212048671) for the first commit (3524ed7) of this PR adding only a test case.

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -22,7 +22,7 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
     public function then(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null)
     {
         if (null !== $this->result) {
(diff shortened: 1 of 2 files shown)


jsor:
Ping @clue

PR #96, opened 2017-08-07
Lock Travis distro so new defaults will not break the build

clue:
Travis is in the process of upgrading the base distro (https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming) and despite all PRs currently being "green", may start to mark the current master as broken. Unlike reactphp/stream#110 this project is not affected by Travis dropping support for older versions of PHP.

Originally from clue/php-connection-manager-extra#24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ php:
   - nightly
   - hhvm
 
+# lock distro so new future defaults will not break the build
+dist: precise
+
 # also test against HHVM, but require "trusty" and ignore errors
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,13 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - nightly
-  - hhvm
+  - nightly # ignore errors, see below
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,8 +6,16 @@ php:
...
+  - nightly # ignore errors, see below
+  - hhvm # ignore errors, see below
+
+# lock distro so new future defaults will not break the build
+dist: trusty
+
+matrix:
+  allow_failures:
+    - php: hhvm
+    - php: nightly
```
jsor: What's the reason for allowing failures for hhvm and nightly now?

clue: The short version is that HHVM builds on Travis repeatedly broke in the past and needed updates despite otherwise working perfectly fine and nightly builds can't really be guaranteed in the future as they have the potential of breaking once PHP 8 or PHP 9 is out.

Also, this is now in line with all other components :+1: 

See also https://github.com/reactphp/socket/pull/109 and https://github.com/reactphp/dns/pull/56

PR #100, opened 2017-09-09
adapt for amphp promises

umbri:
same as in https://github.com/amphp/amp/blob/master/lib/functions.php#L226 for ```\Amp\Promise```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "php": ">=5.4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8"
+        "phpunit/phpunit": "~4.8",
+        "amphp/amp": "^2"
     },
     "autoload": {
         "psr-4": {
--- a/src/functions.php
+++ b/src/functions.php
@@ -234,3 +234,23 @@ function _checkTypehint(callable $callback, $object)
 
     return $expectedException->getClass()->isInstance($object);
 }
+
+/**
+ * @param \Amp\Promise $promise
+ *
+ * @return PromiseInterface
+ */
+function adapt(\Amp\Promise $promise): PromiseInterface
+{
+    $deferred = new Deferred();
+
+    $promise->onResolve(function ($error = null, $result = null) use ($deferred) {
+        if ($error) {
+            $deferred->reject($error);
+        } else {
+            $deferred->resolve($result);
+        }
+    });
+
+    return $deferred->promise();
+}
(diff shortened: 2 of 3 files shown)


cboden:
I think this type of support would be better in its own repo. I don't like the idea of introducing an API that relies on a library that may or may not be available.

umbri:
https://github.com/umbri/amp-react-interop

umbri:
Just in case: I am using ```namespace React\Promise```
https://github.com/umbri/amp-react-interop/blob/master/lib/functions.php#L33 

Should I change that ?

cboden:
Yes please. React is this organizations namespace.

umbri:
Ok, no problem I will change that

kelunik:
@cboden While a separate package could also make sense, I think including something like that into another package like this one directly can make sense for improved discoverability.

The type declaration could also be dropped in the future to allow for other promise implementations to be easily adapted, too.

kelunik:
@umbri I think it would make sense to simply support it in [here](https://github.com/reactphp/promise/blob/84afb3d0b96b58d5429948c83a6330da657d35fe/src/functions.php#L5) instead of a separate `adapt` method.

jsor:
I think it makes perfect sense to provide that via a separate package. If Amp wants to be consumable by promise libs, it can provide a `then()` method. This would make it consumable by [ReactPHP's promises](https://github.com/reactphp/promise/blob/177fe310cc1c0af01fdb03b4dc104ee7c3a2ba19/src/functions.php#L11) and also [Guzzle's promises](https://github.com/guzzle/promises/blob/09e549f5534380c68761260a71f847644d8f65aa/src/functions.php#L73).

clue:
Thanks for the PR @umbri, I think this sparked a very good discussion and lead the right thing: Your adapter is much appreciated and I suppose something that people that are interested in using amphp with ReactPHP would be interested in :+1: 

Other than that, I agree with @cboden and @jsor in that it doesn't make much sense to provide adapters for all kinds of (niche) third party projects here, in particular given that we already support "foreign promises" (anything with a standard `then()` method).

@kelunik As much as I understand your effort, I don't thinks this project should be concerned with "improved discoverability" of (niche) third party projects that this project has no involvement whatsoever with.

I believe this has been answered, so I'm closing this for now. Please come back with more details if this persists and we can reopen this :+1:

Again, thanks for the discussion everybody and thanks @umbri for bringing this idea to the table :+1:

PR #117, opened 2018-05-04
Improve memory consumption by using static child canceller callback without binding to parent promise

clue:
Calling `cancel()` on a child promise that follows its parent promise now no longer causes a cyclic garbage reference in any exception trace as discussed in #46. This builds on top of the implementation approach in #115 and #116 and the ideas discussed in #46.

The gist here is that `$promise->then()->cancel()` now no longer causes any unexpected cyclic garbage reference and consumers of this package do not need to take special care of this.

Similar patches have been introduced with #115 and #116 to implement a similar logic for the resolution and cancellation callbacks and this PR takes advantage of this for its internal implementation.

Invoking the benchmarking example from #113 shows no effect because it does not use child promises. After patching this to use `$promise->then()->cancel()` instead of `$promise->cancel()`, this shows a very significant performance and memory improvement! Initially this peaked somewhere around 15 MB on my system taking 24s. After applying this patch, this script reports a constant memory consumption of around 0.6 MB taking 5s

This PR actually includes a test that shows how garbage memory references are no longer an issue in any supported PHP version and how cancelling a child promise no longer causes any such references on its own (this means that this requires no effort on the consumer side).


src/Promise.php
```diff
@@ -28,15 +28,41 @@ public function then(callable $onFulfilled = null, callable $onRejected = null,
...
-        return new static($this->resolver($onFulfilled, $onRejected, $onProgress), function () {
-            $this->requiredCancelRequests--;
+        // keep a reference to this promise instance for the static canceller function.
+        // see also parentCancellerFunction() for more details.
+        $parent = $this;
+        ++$parent->requiredCancelRequests;
 
-            if ($this->requiredCancelRequests <= 0) {
-                $this->cancel();
+        return new static($this->resolver($onFulfilled, $onRejected, $onProgress), self::parentCancellerFunction($parent));    }
```
jsor: CS: missing linebreak before the closing `}` of the method

clue: Thanks for spotting, fixed! :shipit:

PR #121, opened 2018-06-01
Properly instantiate LazyPromise

holtkamp:


holtkamp:
And while we are at it? Maybe a more (concrete) real life example on how to use this? Tried to understand the example and tried to use it, but I find it hard to understand...

Context: I try to use lazy Promises in a [Queue](https://github.com/clue/reactphp-mq) to reduce the amount of used / allocated memory...

jsor:
@holtkamp The LazyPromise is just a wrapper around a callback which produces the "real" promise only once `then()` is called on the LazyPromise. Since almost in all cases, `then()` is called immediately, it doesn't provide a real benefit and thus we decided to remove in for v3.0 (#98).

holtkamp:
@jsor thanks for the explanation!

PR #136, opened 2019-02-20
Add PHPDoc in Promise::then method

tifabien:
I'm using PHPStan to check my code.
When I'm using chaining method calls with `Promise->then(...)->otherwise(...)`, PHPStan is complaining about the fact that `otherwise` method is undefined in `PromiseInterface`.
I can break my chaining method calls and use the PHPDoc comment but it's not very sexy...
```
/** @var ExtendedPromiseInterface $promise */
$promise = $this->session->call(...)->then(...);
$promise->otherwise(...);
```

By adding the PHPDoc it fixes this warning and I can continue to use chaining method calls.

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -25,6 +25,9 @@ class Promise implements ExtendedPromiseInterface, CancellablePromiseInterface
         $this->call($cb);
     }
 
+    /**
+     * @return static
+     */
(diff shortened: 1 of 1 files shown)


jsor:
Thanks for your PR. Unfortunately, this would be a BC break, see #139 for a discussion around this.

PR #156, opened 2020-01-28
Narrow down docblock type of Deferred::promise

Gert-dev:
The interface indicates it returns a `PromiseInterface`, but the implementation returns a `Promise`, which also implements `ExtendedPromiseInterface`.

Calling `Deferred::promise()` will make PHPStan complain if you need to return an `ExtendedPromiseInterface`, as it cannot deduce that it does indeed return it from the implementation alone.

This should also aid autocompletion in IDEs.

### DIFF
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -15,6 +15,9 @@ class Deferred implements PromisorInterface
         $this->canceller = $canceller;
     }
 
+    /**
+     * @return Promise
+     */
     public function promise()
     {
         if (null === $this->promise) {


Gert-dev:
I didn't consider that, but you are right; it is a subtle BC break.

I see that you indeed added `PromiseInterface` on the `master` branch and the `ExtendedPromiseInterface` doesn't appear to exist anymore there, so this becomes a nonissue there. I'll close this pull request in that case, thanks!

PR #169, opened 2020-06-18
Add compose function

ace411:
Chaining functions is endemic to writing code asynchronously, with promises. In this Pull Request is a simple primitive that allows for users of the library to effortlessly chain functions applicable on values subsumed in Promises.

A sequence like the one shown below presents an alternative to forward chaining with the `then()` method. 

```php
$operations = compose(
  fn ($x) => resolve($x ** 2),
  fn ($x) => $x - 5
);

$operations(resolve(2));
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Table of Contents
      * [some()](#some)
      * [map()](#map)
      * [reduce()](#reduce)
+     * [compose()](#compose)
    * [PromisorInterface](#promisorinterface)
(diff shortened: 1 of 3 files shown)


clue:
I'm having to close this for now as it hasn't received any input in a while and it's unlikely this will get traction any time soon. Please come back with more details if you think this is still relevant and we can reopen this :+1: 

Thank you for your effort nonetheless, keep it up! :+1:

PR #182, opened 2021-01-19
Implement JsonSerializable

simonbuehler:
when beeing part of e.g. array the FullfilledPromise is not serialized, this will return the value now

### DIFF
--- a/src/Internal/FulfilledPromise.php
+++ b/src/Internal/FulfilledPromise.php
@@ -2,6 +2,7 @@
 
 namespace React\Promise\Internal;
 
+use JsonSerializable;
 use React\Promise\Promise;
 use React\Promise\PromiseInterface;
 use function React\Promise\enqueue;
@@ -11,7 +12,7 @@ use function React\Promise\resolve;
 /**
  * @internal
  */
-final class FulfilledPromise implements PromiseInterface
+final class FulfilledPromise implements PromiseInterface , JsonSerializable
 {
     private $value;
 
@@ -76,5 +77,10 @@ final class FulfilledPromise implements PromiseInterface
 
     public function cancel(): void
     {
+    }
+    
+    public function jsonSerialize()
+    {
+        return $this->value;
     }
 }


clue:
Hi @simonbuehler, thanks for taking the time to file this PR! :+1: 

Unfortunately, the PR does not contain any documentation or tests, so it's a bit unclear what you're trying to achieve from the changeset you've posted.

I don't think this piece of code should live in the promise implementation, but I'm curious to get a better understanding of the problem you're trying to solve.

May I ask you provide a simple gist of what you're trying to achieve? Thanks!

simonbuehler:
hi,

sorry for the barebone commit, i just fixed a problem i came across when returning an object in laravel that had a nested FulfilledPromise in it and after serializing the value was lost.

the image shows a laravel dump of the object on serverside  and below is the returned http response after going through the serialisation 



the call to json_encode was the problem i fixed along the way, please let me know what would be a better way to accomplish this

WyriHaximus:
Ok but why is the promise being JSON encoded? When you get a promise you don't know it has resolved yet until your handlers are called. You might just as easy end up with a pending promise and a null value. 

I also agree with @clue that this code has no place in this project, and would even create the wrong idea of what this package does. So I rather focus on understanding what you're trying to do and help you resolve that.

simonbuehler:
in my understanding its the right place as its in the `FulfilledPromise` and there the value should be settled?

> 

> Fulfilled is a state of a Promise. It means that the promise has been resolved and now has its resolved value (using the internal resolve function). The operation represented by the promise has been completed successfully.

simonbuehler:
for the context: im using the LaravelWebSockets server which is based on Ratchet and Reactphp - inside a API handler i prepare the websocket user data together with the sockets along those lines:

``` php
public function __invoke(Request $request)
    {
        return $this->channelManager
            ->getChannelMembers($request->appId, $request->channelName)
            ->then(function ($members) use ($request) {
                $users = collect($members)->map(function ($user) use ($request) {
                    $user->sockets = $this->channelManager->getMemberSockets($user->user_id, $request->appId, $request->channelName)
                        ->then(function ($sockets) {
                            return $sockets; //resolved
                        });
                    return $user; //resolved
                })->values()->toArray();
                dump($users); //resolved - with sockets 
                return [
                    'users' => $users
                ];
            });
    }
```

then the laravel magic returns the result to the http caller in
https://github.com/beyondcode/laravel-websockets/blob/8baefdd4c41c3c9b46b324a2d14a388128682474/src/API/Controller.php#L217 (  `tap($connection)->send(JsonResponse::create($response))->close(); `)

WyriHaximus:
Try this:

```php
public function __invoke(Request $request)
    {
        return $this->channelManager
            ->getChannelMembers($request->appId, $request->channelName)
            ->then(function ($members) use ($request) {
                return \React\Promise\all([
                    'users' => \React\Promise\all(collect($members)->map(function ($user) use ($request) {
                    $this->channelManager->getMemberSockets($user->user_id, $request->appId, $request->channelName)
                        ->then(function ($sockets) use ($user) {
                            $user->sockets = $sockets; //resolved
                            return $user;
                        });
                    return $user; //resolved
                })->values()->toArray())
                ];
            });
    }
```

simonbuehler:
wow 🙌

its works and is so elegant - but i also have to let it sink in a while! Thanks for the input guys!

PR #188, opened 2021-03-15
[3.x] Add template annotations

simPod:
I was considering adapting this lib for promises though it's currently impossible to be type-safe with it. 

I found this psalm plugin https://github.com/Bocmah/psalm-reactphp-promise-plugin that only includes stubs for psalm. So I was thinking it can be simply ported into this promise lib so it's also supported by phpstan and by any static analysis overall without need for any additional setup.

### DIFF
--- a/src/PromiseInterface.php
+++ b/src/PromiseInterface.php
@@ -2,6 +2,10 @@
 
 namespace React\Promise;
 
+/**
+ * @psalm-template T
+ * @psalm-template R
+ */
 interface PromiseInterface
 {
     /**
@@ -28,9 +32,15 @@ interface PromiseInterface
      *  2. `$onFulfilled` and `$onRejected` will never be called more
      *      than once.
      *
-     * @param callable|null $onFulfilled
-     * @param callable|null $onRejected
-     * @return PromiseInterface
+     * @template TFulfilled of mixed
+     * @template TRejected of mixed
+     * @param (callable(T): (PromiseInterface<TFulfilled>|TFulfilled))|null $onFulfilled
+     * @param (callable(R): (PromiseInterface<TRejected>|TRejected))|null $onRejected
+     * @return PromiseInterface<(
+     *   $onFulfilled is not null
+     *     ? ($onRejected is not null ? TFulfilled|TRejected : TFulfilled)
+     *     : ($onRejected is not null ? TRejected : R)
+     * )>
      */
     public function then(?callable $onFulfilled = null, ?callable $onRejected = null): PromiseInterface;
 
@@ -44,8 +54,9 @@ interface PromiseInterface
      * Additionally, you can type hint the `$reason` argument of `$onRejected` to catch
      * only specific errors.
      *
-     * @param callable $onRejected
-     * @return PromiseInterface
+     * @param callable(mixed):(PromiseInterface<T>|mixed) $onRejected
+     *
+     * @return PromiseInterface<T>
      */
     public function catch(callable $onRejected): PromiseInterface;
 
@@ -92,7 +103,7 @@ interface PromiseInterface
      * ```
      *
      * @param callable $onFulfilledOrRejected
-     * @return PromiseInterface
+     * @return PromiseInterface<T>
      */
     public function finally(callable $onFulfilledOrRejected): PromiseInterface;
 
@@ -118,7 +129,7 @@ interface PromiseInterface
      * ```
      *
      * @param callable $onRejected
-     * @return PromiseInterface
+     * @return PromiseInterface<T>
      * @deprecated 3.0.0 Use catch() instead
      * @see self::catch()
      */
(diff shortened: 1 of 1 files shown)


clue (commented):
@simPod Thanks for bringing this up!

I think this feature makes perfect sense, but I'd like to make sure this is complete before merging this.

In particular, can you add some automated tests to verify this works as expected?

On top of this, phpstan also supports generics, does it make sense to address this as part of the PR?

simPod:
@clue thanks for input. I'll look into how to test this.

Also note this is only partial implementation (for happy path) as both psalm and phpstan do not support default template types yet (it's impossible to resolve template to a type when `null` is passed).

This is resolving `onFulfilled`'s type but not `onRejected`'s one. I thought it might be a good start anyway to support at least one side for now. Or we keep it here and build up as the support emerges.

Crosslinks:  
https://github.com/vimeo/psalm/issues/5407  
https://github.com/phpstan/phpstan/issues/4801 

> On top of this, phpstan also supports generics, does it make sense to address this as part of the PR?

phpstan reads `@psalm-` annotations so this should be addressed if you meant this.  
(personally, I don't use prefixes at all in my code but as I ported the initial implementation from psalm plugin, I kept them)

src/PromiseInterface.php
```diff
@@ -44,7 +48,7 @@ public function then(?callable $onFulfilled = null, ?callable $onRejected = null
      * Since the purpose of `done()` is consumption rather than transformation,
      * `done()` always returns `null`.
      *
-     * @param callable|null $onFulfilled
+     * @param callable(mixed):TResolve|null $onFulfilled
```
Seldaek, line 51: I'd think this should be `TResolved`, but it's not defined on this method so I don't think that would work? Not sure what this is supposed to do here.

simPod:
I'll wait for default promise value support in SA since it's fairly useless without it

SimonFrings:
@simPod thanks for your work on this so far :+1: 

To keep you updated:
@WyriHaximus, @clue and I are currently looking into this, we'll have a call later this day to discuss the details. There's also a discussion in pslam with more information on this: [#7559](https://github.com/vimeo/psalm/discussions/7559).

simPod:
Thank you for update and pushing it forward!

WyriHaximus:
One major thing for me was to tackle the bigger picture, so not just this promise package. But also `async` and `await` and observables plus await with observables and get it right as a whole.

WyriHaximus:
@simPod Just a heads up: I'm going to make a bunch of suggested changes based on https://psalm.dev/r/22a9692a97 later today. But want thank you for your work so far on this and for pushing it forward so far :+1:

src/PromiseInterface.php
```diff
@@ -2,6 +2,7 @@
 
 namespace React\Promise;
 
+/** @psalm-template T */
```
WyriHaximus:
```suggestion
/** @template T */
```

src/PromiseInterface.php
```diff
@@ -28,9 +29,12 @@ interface PromiseInterface
...
      *
-     * @param callable|null $onFulfilled
-     * @param callable|null $onRejected
-     * @return PromiseInterface
+     * @psalm-template TResolved
+     *
+     * @param callable(T):(PromiseInterface<TResolved>|TResolved)|null $onFulfilled
+     * @param callable(mixed):(PromiseInterface|mixed)|null $onRejected
+     *
+     * @return PromiseInterface<TResolved>
```
WyriHaximus:
```suggestion
     * @template TReturn of mixed
     * 
     * @param callable(T): TReturn $callable
     * 
     * @return (TReturn is PromiseInterface ? TReturn : PromiseInterface<TReturn>)
```

src/PromisorInterface.php
```diff
@@ -2,12 +2,13 @@
 
 namespace React\Promise;
 
+/** @psalm-template T */
```
WyriHaximus:
```suggestion
/** @template T */
```

WyriHaximus (changes requested):
When making the suggestions, I realized I forgot about the other methods, will update https://github.com/vimeo/psalm/discussions/7559 and this PR's suggestions soon

WyriHaximus:
FYI just put up #223 as a draft for the 2.x branch so I can pull it in through composer and test it out again in a project without having to manually edit files again. Will try to get back shortly and hope live doesn't get in between again

src/PromiseInterface.php
```diff
@@ -2,6 +2,10 @@
 
 namespace React\Promise;
 
+/**
+ * @psalm-template T
+ * @psalm-template R
```
spawnia, line 7: Any reason this can not just use `@template`?

WyriHaximus:
FYI just filed https://github.com/reactphp/promise/pull/227 and https://github.com/reactphp/promise/pull/228 those will not have any direct affect on this PR as it's for the upcoming v3 release. We want basic support out there and then work on getting near to a 100% support out there through this PR and PR's for 1.x and 2.x.

WyriHaximus:
@simPod Hey if you don't mind I'm going to use your commit as a base to get template annotations + some additional type insurance in.

simPod:
Awesome 👌🏿

WyriHaximus:
PR is up at: https://github.com/reactphp/promise/pull/247

PR #193, opened 2021-08-31
Reduced promises depth test

woodongwong:
When I execute phpunit, it prompts me `Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes)`, which I know I can fix by modifying php.ini. I think I can reduce the promises depth, i.e. I can not modify the memory_limit (default 128M) and also ensure the correct test, thanks. (very sorry, English is not my native language)

### DIFF
--- a/tests/FunctionResolveTest.php
+++ b/tests/FunctionResolveTest.php
@@ -130,12 +130,12 @@ class FunctionResolveTest extends TestCase
     {
         $deferreds = [];
 
-        for ($i = 0; $i < 250; $i++) {
+        for ($i = 0; $i < 150; $i++) {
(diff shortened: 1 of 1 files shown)


SimonFrings:
@woodongwong thanks for looking into this :+1:
You're right with the `memory_limit` being 128M by default, so i don't see any reason to reject your changes.

For now you can raise your `memory_limit` by executing the test suite like this:
```
$ php -d memory_limit=256M vendor/bin/phpunit
```

SimonFrings (approved):

jsor (approved):

PR #196, opened 2021-09-28
Run CI on PHP8.0

bzikarsky:
No explanation needed


bzikarsky:
The second commit forces PHPunit to `E_ALL`. This uncovers one deprecation error. A fix for this errors can be found in !195

PR #202, opened 2021-11-19
ExtendedPromiseInterface: type-hint for then()

Thomas-Gelf:
This solves the problem that a PHP IDE always believes your then()
method returns a PromiseInterface and not an ExtendedPromiseInterface

### DIFF
--- a/src/ExtendedPromiseInterface.php
+++ b/src/ExtendedPromiseInterface.php
@@ -2,6 +2,9 @@
 
 namespace React\Promise;
 
+/**
+ * @method ExtendedPromiseInterface then(callable $onFulfilled = null, callable $onRejected = null, callable $onProgress = null)
+ */
 interface ExtendedPromiseInterface extends PromiseInterface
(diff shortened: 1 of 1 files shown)


SimonFrings:
@Thomas-Gelf thanks for looking into this :+1: 

It's not 100% guaranteed that this will always return a ExtendedPromiseInterface. I discussed this with @clue and maybe there should be something like `@return PromiseInterface || ExtendedPromiseInterface` inside the docblock  of the `then()` method inside `PromiseInterface`.
This should lead to the same results as your addition.

What do you think about this?

SimonFrings:
This pull request here shouldn't be needed anymore with Promise v3 out, as we added template types and improved type safety in general in #246, #247 and others. The functionality of the `ExtendedPromiseInterface` is now part of `PromiseInterface`  and has been removed in #75.

This is why I'll go ahead and close this ticket here, thanks for the contribution :+1:

PR #212, opened 2022-02-11
Rename development branch to `3.x` and add installation instructions

clue:
This changeset suggests renaming the development branch to `3.x` and adds appropriate installation instructions. In particular, this makes it much easier to install the v3 development version or any version like this:

```bash
$ composer require react/promise:^3@dev
$ composer require react/promise:"^3@dev || ^2 || ^1"
```

This has originally been brought up in #204 by @Seldaek, but instead of defining a branch alias, this PR suggests simply renaming what is currently `master` to `3.x`. This is in line with our existing `2.x` and `1.x` branches and means we do not have to update any alias definitions once we would start working on a potential future `4.x`.

From a consumer's perspective, this means installation is now much easier, as targeting `^3@dev` (or `3.x-dev`) is semantically more obvious and safer than using `dev-master`. Additionally, keep in mind that branch names are development artifacts that are subject to change at some point in the future. Once we would decide to rename or delete any of our version branches, the `^3@dev` reference would still be valid as it would also match stable versions if we have any `^3` tags at this point (which is safe to assume at this point).

Once this PR is merged, I will manually rename the `master` branch to `3.x`. At the same time, I also propose to set the current default branch to `2.x` until a stable 3.0.0 version has been released. This is more consistent with our website, which also renders the latest tagged release (2.x) instead of the latest development version (3.x). Once a stable 3.0.0 version is released, we should set the default branch to `3.x`.

Also refs https://github.com/reactphp/async/pull/14 and https://github.com/reactphp/async/pull/11 for similar branch names used in react/async.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@ A lightweight implementation of
 
 [![CI status](https://github.com/reactphp/promise/workflows/CI/badge.svg)](https://github.com/reactphp/promise/actions)
 
-> The master branch contains the code for the upcoming 3.0 release.
-> For the code of the current stable 2.x release, checkout the 
-> [2.x branch](https://github.com/reactphp/promise/tree/2.x).
-
-> The upcoming 3.0 release will be the way forward for this package. 
-> However we will still actively support 2.0 and 1.0 for those not yet 
-> on PHP 7+.
+> **Development version:** This branch contains the code for the upcoming 3.0 release.
+> For the code of the current stable 2.x release, check out the
+> [`2.x` branch](https://github.com/reactphp/promise/tree/2.x).
+>
+> The upcoming 3.0 release will be the way forward for this package.
+> However, we will still actively support 2.0 and 1.0 for those not yet
+> on PHP 7.1+.
 
 Table of Contents
 -----------------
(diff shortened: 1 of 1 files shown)


SimonFrings (approved):

jsor (approved):

WyriHaximus (approved):

clue:
> Once this PR is merged, I will manually rename the `master` branch to `3.x`. At the same time, I also propose to set the current default branch to `2.x` until a stable 3.0.0 version has been released. This is more consistent with our website, which also renders the latest tagged release (2.x) instead of the latest development version (3.x). Once a stable 3.0.0 version is released, we should set the default branch to `3.x`.

clue:
> Once this PR is merged, I will manually rename the `master` branch to `3.x`. At the same time, I also propose to set the current default branch to `2.x` until a stable 3.0.0 version has been released. This is more consistent with our website, which also renders the latest tagged release (2.x) instead of the latest development version (3.x). Once a stable 3.0.0 version is released, we should set the default branch to `3.x`.

After discussing this matter again with @WyriHaximus and @SimonFrings last week, we've come to the conclusion that `3.x` should be the default branch instead of `2.x` to give more visibility to the current development. This should not otherwise affect development or any pending PRs but would be more consistent across the entire project, in particular with regards to the plans for the upcoming v3 as discussed in https://github.com/orgs/reactphp/discussions/472. I've updated the repository accordingly to select `3.x` as a default branch.

PR #217, opened 2022-04-12
[2.x] Fix legacy HHVM build by downgrading Composer

SimonFrings:
Builds on top of the work done by @clue and the discussion inside reactphp/socket#289.

I also updated the path for the CI badge.


clue (approved):
@SimonFrings Thanks for applying the changes here as well! :shipit:

Also good to see the same changes backported to v1 via #218 and especially the fact this isn't needed for Promise v3 due to #138! :+1:

WyriHaximus (approved):

Issue #35, opened 2015-07-06
Validate array argument for promise-array related functions


jsor:
Reject all() / race() / any() / some() / map() / reduce() when called with a non-array.

```php
all(null)
    ->otherwise(function(\InvalidArgumentException $e) {
        assertEquals('Expecting an array or a promise which resolves to an array, got NULL instead.' , $exception->getMessage());
    });
```

clue:
:+1: for rejecting invalid input

But then again *what* is valid input here? :-)

jsor:
At the moment, it is an array or a promise which resolve to an array. Passing anything other produces

```php
all(null)
    ->then(function($value) {
        assertEquals([], $value);
    });
```

which i think is not an expected behaviour.

clue:
> which i think is not an expected behaviour

Yeah, I agree that we should handle this is in a more sane way :+1: 

However, now that we're already considering *what* is valid input, I would suggest using a stricter definition. For example, ES6-style promises limit the parameter to be an array of promises (or values which are considered like resolved promises): https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Promise/all (which IMHO makes sense to me).

As such, does it make sense to pass a single Promise which resolves to an array of values? Do we have a valid use case for this?

jsor:
> As such, does it make sense to pass a single Promise which resolves to an array of values? Do we have a valid use case for this?

I've never used a single input promise.

Options:
* No change, allow input promise which resolves to an array, requires rejection with InvalidArgumentException for non-array input
* Typehint against `array`, does **not** require rejection for non-array input
* Allow `array` **and** `\Traversable` input, no typehint possible, requires rejection with InvalidArgumentException for non-array input

I'm unsure on this atm.

clue:
> No change, allow input promise which resolves to an array, requires rejection with InvalidArgumentException for non-array input

:-1: on this unless we find a valid use case

> *    Typehint against array, does not require rejection for non-array input
*    Allow array and \Traversable input, no typehint possible, requires rejection with InvalidArgumentException for non-array input

IMO both sound sane…

Some (random) thoughts:

* On success, the `all()` method resolves with an `array`, so for consistency it may make sense to require an input `array`
* `Traversable` and `Iterator` have valid use cases, though I'm failing to see how these would be beneficial here. Also, converting via `iterator_to_array()` is trivial.
[...]

PR #219, opened 2022-05-17
[RFC] Remove `some()`, `map()`, `reduce()` functions

clue:
This simple PR suggests removing the `some()`, `map()`, `reduce()` functions to reduce the API surface. With these changes applied, our API is pretty much identical to ES6 promises commonly used in JavaScript. This makes our APIs appear less complex and thus easier to understand.

Empirical evidence suggests these functions aren't used very commonly in our ecosystem either, so I'd rather use the chance to clean up our API surface with the upcoming Promise v3 release. I don't currently see a prominent need for these functions, but I'm open to the idea of reintroducing them later if needed. There's reason to believe that these functions could (should?) exist as part of our new reactphp/async package, especially considering how it mimics part of async.js.

I'm filing this as an RFC to get more feedback on these functions and to see how others feel about this. If this gets merged, I'll file a follow-up PR to deprecate these functions for Promise v2 to ease upgrading.

Refs #35


jsor (approved):

clue:
Rebased to resolve merge conflict now that #220 is in :shipit:

WyriHaximus (approved):
:+1:

In case questions do come up, we can consider putting them in a @friends-of-reactphp package.

PR #220, opened 2022-05-17
[RFC] Remove unneeded `PromisorInterface`

clue:
This simple changeset suggests removing the unneeded `PromisorInterface` (not to be confused with `PromiseInterface` which remains unchanged!). This is done in an attempt to reduce the API surface and make the API less confusing and easier to understand. On top of this, this means our APIs are closer to ES6 promises commonly used in JavaScript (see also #219).

In particular, its description currently reads "provides a common interface for objects that provide a promise". While technically correct, I consider this to be misleading as most "objects that provide a promise" as a return value (think an async `connect()` or `fetch()` method) would not implement this interface at all.

Empirical evidence suggests this interface isn't used much in our ecosystem either, so I'd rather use the chance to clean up our API surface with the upcoming Promise v3 release.

Like #219, I'm filing this as an RFC to get more feedback on this interface and to see how others feel about this. If this gets merged, I'll file a follow-up PR to deprecate this interface for Promise v2 to ease upgrading.


jsor (approved):

WyriHaximus (approved):
Honestly? I never ever used this, noir will I ever do :+1:

PR #236, opened 2022-09-08
[3.x] feat: add support for DNF types

### LINKED ISSUE #235: Support Disjunctive Normal Form Types (DNF types) for PHP 8.2+
We should support Disjunctive Normal Form Types (DNF types) for PHP 8.2+ (RFC: https://wiki.php.net/rfc/dnf_types):

```php
$promise->otherwise(function ((A&B)|(C&D) $e) { }); // Promise v2
$promise->catch(function ((A&B)|(C&D) $e) { }); // Promise v3
```

* Support for union types (PHP 8+) has been added via #198 (Promise v2) and #197 (Promise v3)
* Support for intersection types (PHP 8.1+) has been added via #195 (Promise v2) and #209 (Promise v3)
* The `catch()` method exists for Promise v3 only, the `otherwise()` method has been deprecated for Promise v3 via #208
* There's currently no support for DNF types in either Promise v2 or Promise v3. Support should be added by adding this feature to either version (preferable Promise v3) and then cherry-picking to the other version.
* Legacy Promise v1 is not affected as it does not have an `otherwise()` or `catch()` method that checks exception types.
* DNF types will be supported in PHP 8.2+ only which is scheduled to be released in November.
* Other than that, this project supports PHP 8.2+ just fine, see #230 and #234

Input and PRs very much welcome! :+1:

nhedger:
This PR adds support for Disjunctive Normal Form Types (DNF types) when using PHP 8.2.

### Overview of the changes

- [x] Removes the previous guard that would only allow `ReflectionNamedType` to be analyzed.
- [x] ~Adds two new **internal** functions `matchesType` and `matchesAllTypes` to extract common functionality.~
- [x] Adds tests + supporting fixtures

Please let me know if you think that I missed something.

---

Closes #235

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -267,14 +267,12 @@ function _checkTypehint(callable $callback, \Throwable $reason): bool
     }
 
     foreach ($types as $type) {
-        if (!$type instanceof \ReflectionNamedType) {
-            throw new \LogicException('This implementation does not support groups of intersection or union types');
+
+        if ($type instanceof \ReflectionIntersectionType) {
+            $matches = matchesAllTypes($reason, $type->getTypes());
+        } else {
+            $matches = matchesType($reason, $type);
         }
-
-        // A named-type can be either a class-name or a built-in type like string, int, array, etc.
-        $matches = ($type->isBuiltin() && \gettype($reason) === $type->getName())
-            || (new \ReflectionClass($type->getName()))->isInstance($reason);
-
 
         // If we look for a single match (union), we can return early on match
         // If we look for a full match (intersection), we can return early on mismatch
@@ -292,4 +290,27 @@ function _checkTypehint(callable $callback, \Throwable $reason): bool
     // If we look for a single match (union) and did not return early, we matched no type and are false
     // If we look for a full match (intersection) and did not return early, we matched all types and are true
     return $isTypeUnion ? false : true;
+}
+
+/**
+ * @internal
+ */
+function matchesType($reason, $type): bool
+{
+    return ($type->isBuiltin() && \gettype($reason) === $type->getName())
+        || (new \ReflectionClass($type->getName()))->isInstance($reason);
+}
+
+/**
+ * @internal
+ */
+function matchesAllTypes($reason, $types): bool
+{
+    foreach ($types as $typeToMatch) {
+        if (!matchesType($reason, $typeToMatch)) {
+            return false ;
+        }
+    }
+
+    return true;
 }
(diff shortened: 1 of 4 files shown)


SimonFrings (approved):
@nhedger Looks good to me 👍

src/functions.php
```diff
@@ -293,3 +291,26 @@ function _checkTypehint(callable $callback, \Throwable $reason): bool
     // If we look for a full match (intersection) and did not return early, we matched all types and are true
     return $isTypeUnion ? false : true;
 }
+
+/**
+ * @internal
+ */
+function matchesType($reason, $type): bool
```
clue: I see why this was done, but not a big fan of `@internal` functions here as they still clutter the namespace and may interfere with autocompletion. I consider this more a matter of preference, but do you think it's reasonable to inline the function definition considering it's not referenced anywhere else?

src/functions.php
```diff
@@ -293,3 +291,26 @@ function _checkTypehint(callable $callback, \Throwable $reason): bool
...
+{
+    return ($type->isBuiltin() && \gettype($reason) === $type->getName())
+        || (new \ReflectionClass($type->getName()))->isInstance($reason);
+}
+
+/**
+ * @internal
+ */
+function matchesAllTypes($reason, $types): bool
```
clue: This function and the one above do not have type declarations. See note above, perhaps we don't need these functions in the first place?

clue (commented):
@nhedger Thank you very much for these high quality changes! Functionally, this LGTM and very much appreciate the added tests.

I've added some minor remarks regarding internal functions only, WDYT?

src/functions.php
nhedger: Hey, I extracted these two functions because `matchesAllTypes` calls `matchesType` and wanted to avoid code duplication. But if you're ok with a bit of duplication to prevent polluting the global namespace, I'll inline them.

clue: Agree it's a tradeoff, so I guess use your own best judgement? Personally, I tend to accept some duplication as long as it's to a reasonable degree and does not repeat any business decisions (DRY).

nhedger: Done

clue (approved):
@nhedger Thanks for the quick update, changes LGTM! Keep it up! :+1:

WyriHaximus (approved):
Awesome, thank you for putting this together :+1:

clue:
@nhedger Excellent work! Ticket #235 suggests applying the same changes also for `2.x`, do you plan to backport this to the other branch or should anybody else take a look at this? :+1:

nhedger:
I'll do it, probably before the end of the week.

PR #237, opened 2022-09-16
[2.x] feat: add support for DNF types on PHP 8.2

### LINKED ISSUE #235: Support Disjunctive Normal Form Types (DNF types) for PHP 8.2+

nhedger:
This PR backports the feature introduced in #236 to `2.x`.

Closes #235

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -354,7 +354,7 @@ function _checkTypehint(callable $callback, $object)
 
     // Extract the type of the argument and handle different possibilities
     $type = $expectedException->getType();
-    
+
     $isTypeUnion = true;
     $types = [];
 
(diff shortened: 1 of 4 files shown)


WyriHaximus (approved):

clue (approved):
@nhedger Thank you for backporting this, changes LGTM! :shipit:

PR #238, opened 2022-10-06
Remove internal `Queue` leftovers

clue:
This changeset removes any internal `Queue` leftovers. This does not change outside behavior of this project and only cleans up unneeded code, so this does not break BC.

This effectively reverts the remainder of the iterative logic introduced in #28/#82/#86. This logic has turned out to cause a number of problems in the past, so the behavior has been reverted as part of #229 already. This means that the iterative approach originally planned for Promise v3 is now reverted completely and the logic is in line with how Promise v2 and v1 behave. The test suite confirms my changes do not break any of the previous assumptions, so this should be safe to apply.

On top of this, this avoids a number of function calls and shows a significant performance improvement in synthetic benchmarks. For example, the following benchmark improved from `16.4s` to ~~`11.6s`~~ `3.1s` on my machine (best of 5 each):

```php
$n = 10_000_000;
for ($i = 0; $i < $n; ++$i) {
    React\Promise\resolve($i)->then(function () { });
}
```

I still agree that an iterative approach sounds *tempting*, but at this point it's unclear what benefit this would actually bring and if it's worth breaking any of the existing behavior. I would suggest we should look into this iterative logic and its consequences again in a follow-up PR some time in the future if we see a need for this. Until we can come up with more specific test cases that highlight now this iterative logic is needed, I would argue that merging this as is to remove any unneeded code and to reduce our complexity seems reasonable.


clue:
Updated to use `resolve()` instead of `new Promise(…)` for wrapping the results as done prior to #28. Again this does not break BC, but significantly improves the benchmark results from `16.4` to `11.6s` to `3.1s`. The benchmark for `3.x` is now pretty much in line with `2.x` again.

SimonFrings (approved):

WyriHaximus (approved):

PR #245, opened 2023-01-18
some() implementation

BitpostNetwork:
Adds a function some() that resolves similarly to all(), but continues to run despite rejecting any of the promises. A new promise is rejected only when all of the promises are rejected.


SimonFrings:
@BitpostNetwork Thanks for filing this PR, we're always happy about contributions :+1:

Can you also add tests so we can assure that your suggested function works as expected.

PR #251, opened 2023-08-30
Update PHPStan

ondrejmirtes:
PHPStan 1.10.31 introduces NonAcceptingNeverType (https://github.com/phpstan/phpstan/issues/6485, https://github.com/phpstan/phpstan/issues/9133) that ensures that a literal `never` type in the code does not accept any other type except `never`.

As a side-effect of this change a type description or an error message about this changed from `*NEVER*` to `never`. I'm sending an update to your test suite because of this :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "php": ">=7.1.0"
     },
     "require-dev": {
-        "phpstan/phpstan": "1.10.20 || 1.4.10",
+        "phpstan/phpstan": "1.10.34 || 1.4.10",
         "phpunit/phpunit": "^9.5 || ^7.5"
     },
     "autoload": {
--- a/tests/types/reject.php
+++ b/tests/types/reject.php
@@ -5,9 +5,9 @@ use function PHPStan\Testing\assertType;
 use function React\Promise\reject;
 use function React\Promise\resolve;
 
-assertType('React\Promise\PromiseInterface<*NEVER*>', reject(new RuntimeException()));
-assertType('React\Promise\PromiseInterface<*NEVER*>', reject(new RuntimeException())->then(null, null));
-// assertType('React\Promise\PromiseInterface<*NEVER*>', reject(new RuntimeException())->then(function (): int {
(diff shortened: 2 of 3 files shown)


clue:
@ondrejmirtes Thank you for looking into this and suggesting improvements for the work done in #247. Briefly eyeballed your changes and noticed there are still a couple of `*NEVER*` types in `tests/types/resolve.php`, can you help me understand the reason for what currently looks like an inconsistency?

ondrejmirtes:
The code in question:

```php
assertType('React\Promise\PromiseInterface<*NEVER*>', resolve(true)->then(function (bool $value): never {
    throw new \RuntimeException();
}));
```

There are two "never" types in PHPStan:

* NeverType (`*NEVER*`) - used basically everywhere - in cases like empty arrays
* NonAcceptingNeverType (`never`) - used when the user types `never` literally in PHPDoc or native return type

For anonymous functions, PHPStan can infer always-throwing function even without a native `never` return type. In that case, `*NEVER*` is used. But there's probably isn't a good reason for that, so I changed it: https://github.com/phpstan/phpstan-src/commit/3e03e9d7c15fdf64e6d55539a6c3045b3b8c6604

It will be released at the earliest convenience.

clue:
@ondrejmirtes Thanks for the insights and the update in PHPStan! :+1:

Your explanation sounds reasonable. It's my understanding this is more of a maintenance topic that doesn't currently have high priority, so we probably want to keep this open until the next PHPStan is released and we can update this consistently in a single shot. Thank you, will keep an eye on this!

WyriHaximus (approved):
Thanks @ondrejmirtes for filing this, had a quick row with this while working on the initial templates implementation. Liking these changes :+1: .

ondrejmirtes:
Just update PHPStan version and all `*NEVER*` to `never`. This should be good to go :)

SimonFrings (approved):
:+1:

clue (approved):
@ondrejmirtes Thanks for the update, changes LGTM! :shipit:

PR #261, opened 2024-05-22
Update test suite to improve PHP 8.4+ support

SimonFrings:
This pull request updates hour test suite to improve support with the upcoming PHP 8.4 release. In particular, the new version breaks BC as it changes the stack trace, which we test for in some of our `phpt` files. 

The PHP 8.4 version is currently still in development, hence I didn't add this to our test matrix to avoid any additional PHP 8.4 feature release breaking something. I explained more about this in #260. I can assure that I have tested this with PHP 8.4 and all tests run successfully with my suggested change.

Builds on top of #248, #258, #259, #260 and others

### DIFF
--- a/tests/FunctionAnyTestRejectedShouldReportUnhandled.phpt
+++ b/tests/FunctionAnyTestRejectedShouldReportUnhandled.phpt
@@ -20,7 +20,7 @@ any([
 --EXPECTF--
 Unhandled promise rejection with React\Promise\Exception\CompositeException: All promises rejected. in %s:%d
 Stack trace:
-#0 %s/src/Promise.php(%d): React\Promise\{closure}(%S)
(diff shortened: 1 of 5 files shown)


SimonFrings:
To be honest, I'm nor quite sure if this is the best way to go about this. I looked for ways to test against all PHP versions in the same `phpt` file, but didn't really find anything. I've noticed that there are already some `phpt` files specifically testing against PHP 7 and PHP 8 in separate files, so I chose this solution for now. Happy about your input on this.

SimonFrings:
Seems like there's no need for additional files afterall, thanks @clue for taking a look at this with me :+1:

clue (approved):
@SimonFrings Thanks for the update, changes LGTM now! :shipit:

WyriHaximus (approved):

PR #262, opened 2024-05-23
Include previous exceptions when reporting unhandled promise rejections

clue:
This changeset makes sure we include any previous exceptions when reporting unhandled promise rejections. In particular, instead of reporting a custom message, we now use PHP's default [`Throwable::__toString()`](https://www.php.net/manual/en/throwable.tostring.php) logic which includes the exception type, its message, stack trace and any previous exceptions by default (https://3v4l.org/lBRM1).

Including any previous exceptions makes sense as a default behavior to be in line with PHP's default exception output and to make it easier to track down any unhandled promise rejections. Like previously, this behavior only triggers for any unhandled promise rejections and can be customized by using a global rejection handler as explained in https://github.com/reactphp/promise#set_rejection_handler.

Builds on top of #248 and #249

### DIFF
--- a/src/Internal/RejectedPromise.php
+++ b/src/Internal/RejectedPromise.php
@@ -37,8 +37,7 @@ final class RejectedPromise implements PromiseInterface
 
         $handler = set_rejection_handler(null);
         if ($handler === null) {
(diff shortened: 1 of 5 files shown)


SimonFrings (approved):
Nice :+1:

WyriHaximus (approved):

PR #255, opened 2023-10-30
[3.x] Run tests on PHP 8.3 and update test suite

clue:
Builds on top of #240 and https://github.com/reactphp/stream/pull/172

PR #268, opened 2025-08-06
Run tests on PHP 8.4 and update test environment

PaulRotmann:
Builds on top of #267, #261, #255 and others.

⚠️ This PR can only be merged after PR #267 has been merged, as it requires the PHP 8.4 compatibility fixes and dependency updates to pass successfully.

Also this PR resolves #266.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,11 @@ on:
 jobs:
   PHPUnit:
     name: PHPUnit (PHP ${{ matrix.php }})
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         php:
+          - 8.4
           - 8.3
           - 8.2
           - 8.1
@@ -34,10 +35,11 @@ jobs:
 
   PHPStan:
     name: PHPStan (PHP ${{ matrix.php }})
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         php:
+          - 8.4
           - 8.3
           - 8.2
           - 8.1


clue (changes requested):
@PaulRotmann Thanks for looking into this, changes look reasonable :+1: 

However, besides the PHPstan update discussed in #267, the test suite currently reports one error on PHP 8.4+ only:

```
There was 1 failure:

1) /home/runner/work/promise/promise/tests/FunctionSetRejectionHandlerThatTriggersDefaultHandlerShouldTerminateProgramForUnhandled.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
-Fatal error: Unexpected RuntimeException: foo in %s line %d
+Deprecated: Passing E_USER_ERROR to trigger_error() is deprecated since 8.4, throw an exception or call exit with a string message instead in /home/runner/work/promise/promise/tests/FunctionSetRejectionHandlerThatTriggersDefaultHandlerShouldTerminateProgramForUnhandled.php on line 9
+
+Fatal error: Unexpected RuntimeException: foo in /home/runner/work/promise/promise/tests/FunctionSetRejectionHandlerThatTriggersDefaultHandlerShouldTerminateProgramForUnhandled.php on line 9

/home/runner/work/promise/promise/tests/FunctionSetRejectionHandlerThatTriggersDefaultHandlerShouldTerminateProgramForUnhandled.phpt:24
```

Here's some background info:

> Calling [trigger_error()](https://www.php.net/manual/en/function.trigger-error.php) with error_level being equal to [E_USER_ERROR](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-user-error) is now deprecated.
>
> — https://www.php.net/manual/en/migration84.deprecated.php

This only affects a single test that specifically tests how the rejection logic behaves when a custom function using this (deprecated) functionality is used. Accordingly, we should probably skip the affected test on PHP 8.4+ (originally introduced in #249):

https://github.com/reactphp/promise/blob/5f80055cc21ba7bcd3989e4902061fc12e2bcc1d/tests/FunctionSetRejectionHandlerThatTriggersDefaultHandlerShouldTerminateProgramForUnhandled.phpt#L15

